PR #4, opened 2024-07-03
feat: Add JSONLanguage#visitorKeys

nzakas:
#### Prerequisites checklist

Added `JSONLanguage#visitorKeys`.

- Added `JSONLanguage#visitorKeys`
- Added tests to ensure that `JSONLanguage#visitorKeys` is properly created

Please double check that the following line is included in the merge commit body to ensure we reset the package version number:

Release-As: 0.1.0

### DIFF
--- a/src/languages/json-language.js
+++ b/src/languages/json-language.js
@@ -9,6 +9,7 @@
 
 import { parse } from "@humanwhocodes/momoa";
 import { JSONSourceCode } from "./json-source-code.js";
+import { visitorKeys } from "@humanwhocodes/momoa";
 
 //-----------------------------------------------------------------------------
 // Types
@@ -58,6 +59,12 @@ export class JSONLanguage {
 	 * @type {"json"|"jsonc"}
 	 */
 	#mode = "json";
+
+	/**
+	 * The visitor keys.
+	 * @type {Record<string, string[]>}
+	 */
+	visitorKeys = Object.fromEntries([...visitorKeys]);
 
 	/**
 	 * Creates a new instance.
--- a/package.json
+++ b/package.json
@@ -60,6 +60,9 @@
     "linting"
   ],
   "license": "Apache-2.0",
+  "dependencies": {
+    "@humanwhocodes/momoa": "^3.1.1"
+  },
   "devDependencies": {
     "@eslint/core": "^0.1.0",
     "@types/eslint": "^8.56.10",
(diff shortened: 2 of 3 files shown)


src/languages/json-language.js
```diff
@@ -9,6 +9,7 @@
 
 import { parse } from "@humanwhocodes/momoa";
 import { JSONSourceCode } from "./json-source-code.js";
+import { visitorKeys } from "@humanwhocodes/momoa";
```
mdjermanovic, line 12: We should update the dependency to `^3.1.1` in package.json.

mdjermanovic:
When I tested this, it turned out that things for which this matters don't actually work (e.g., selectors like `Element + Element`) since ESLint is always passing an empty ancestry to esquery because nodes don't have a `parent` property which [NodeEventGenerator expects here](https://github.com/eslint/eslint/blob/1381394a75b5902ce588455765a3919e2f138a7a/lib/linter/node-event-generator.js#L337-L339). I'll investigate further how this can be fixed in ESLint.

nzakas:
Ah interesting. Yeah, I don't think we can assume every AST type will have a `parent` property, so we probably need to use the `SourceCode#getAncestry()` method.

src/languages/json-language.js
nzakas: This is so weird. I know I updated this locally and tested, and this is the second time you've pointed out that the version hasn't been updated in `package.json`. 🤔

nzakas: Actually, it is updated in `package.json` but for some reason git isn't included it in the branch.

mdjermanovic: It works if `3.1.1` is locally installed, which in most cases would be the case after doing `npm install` since it's the latest in-range version, but to be sure it should be set in package.json.

mdjermanovic (approved):
LGTM, thanks!

mdjermanovic:
> Please double check that the following line is included in the merge commit body to ensure we reset the package version number:
> 
> Release-As: 0.1.0

For some reason, in this repo GitHub doesn't pull anything from the commits into the default commit body.

image

mdjermanovic:
It's probably because we have this setting:

image

Was it intentional? In other repos, we have this:

image

nzakas:
Not intentional. Need to add that to the [new repo template](https://github.com/eslint/team/blob/main/.github/ISSUE_TEMPLATE/new-repo.md).

mdjermanovic:
Merging now as this change is correct for this package. I am working on fixing `NodeEventGenerator` in https://github.com/eslint/eslint/pull/18653.

PR #140, opened 2025-09-13
refactor: use `defaultOptions` in `no-unnormalized-keys`

lumirlumir:
#### Prerequisites checklist

In this PR, I've updated the `no-unnormlized-keys` rule to use `defaultOptions`.

Previously, the rule used custom logic to extract its default value.

According to MDN and `@types/node`, the default value for `String.prototype.normalize` is `"NFC"`, so I've set it as the default.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#parameters

image

image

I've updated the `no-unnormlized-keys` rule to use `defaultOptions`.

#### Related Issues

N/A

N/A

### DIFF
--- a/src/rules/no-unnormalized-keys.js
+++ b/src/rules/no-unnormalized-keys.js
@@ -45,12 +45,16 @@ const rule = {
 				additionalProperties: false,
 			},
 		],
+
+		defaultOptions: [
+			{
+				form: "NFC",
+			},
+		],
 	},
 
 	create(context) {
-		const form = context.options.length
-			? context.options[0].form
-			: undefined;
+		const [{ form }] = context.options;
 
 		return {
 			Member(node) {


lumirlumir:
Hmm, I'm not sure about this 🤔

But since the `sort-keys` rule is already using `defaultOptions` as shown below, and `@eslint/json` is still in the 0.X unstable version, I think it should be safe. Also, we don't have `peerDependencies` in the `package.json` specifying the relevant `eslint` version, so maybe it's fine?

https://github.com/eslint/json/blob/76cd64761477136dab4bb6ec574b2889b19264d3/src/rules/sort-keys.js#L99-L107

aladdin-add:
IMHO, the safest approach is to avoid using `defaultOptions` - we can update these in next major (dropping eslint <9.15).

fasttime:
> maybe we could just update the docs now.

Yes, at a minimum the docs should be updated.

PR #185, opened 2025-11-10
chore: restore eslint-plugin-jsdoc rules

Pixel998:
#### Prerequisites checklist

This PR restores the eslint-plugin-jsdoc rules that were previously disabled, so that JavaScript files once again receive the intended documentation linting behavior from eslint-config-eslint.

Removed the blanket disable of jsdoc rules and cleaned up the configuration structure in `eslint.config.js`.


src/rules/sort-keys.js
```diff
@@ -41,36 +39,60 @@ const hasNonWhitespace = /\S/u;
 const comparators = {
```
DMartens: I think to circumvent, which are in my opinion, useless comments, we could add the type to the variable instead of each function with:
`Record<'ascending' | 'descending', Record<'alphanumeric' | 'natural', Record<'sensitive' | 'insensitive', Comparator>>>`. The type of the keys (e.g. `'alphanumeric' | 'natural'` could be their own type and be used for the converted option variables (e.g. `sortName` on line 166)

DMartens (dismissed):
LGMT, thanks. I left one non-blocking suggestion and leave this open for a day if someone else wants to review.

PR #224, opened 2026-03-25
refactor: update TypeScript to v6

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

## What is the purpose of this pull request?

This PR updates TypeScript to v6.

The following two changes mainly affected us during the upgrade to TypeScript v6.

- `strict` mode is now the default.

https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#simple-default-changes

image

- `rootdir` now defaults to `.`

https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#rootdir-now-defaults-to-.

image

I've tried to retain strict mode behavior as much as possible and only disabled it where it produced too many errors.

### `src/index.js`

The logic below produced a type error after upgrading to TypeScript v6, so I used an object getter instead to avoid the error and to avoid the block-syntax workaround.

https://github.com/eslint/json/blob/c9e816f58848ae37819dae7132b02110e61bd250/src/index.js#L39-L42

Freezing the configuration object wasn't considered a breaking change in https://github.com/eslint/eslint/pull/20116, so using a getter here seems fine, but I'd welcome other opinions.

I've also verified that the output `index.d.ts` file is the same as before.

### `tests/types/tsconfig.json`

- Re-enabled `strictNullChecks` and `useUnknownInCatchVariables` after they were turned off in root level `tsconfig.json`.

### `package.json`

- Added `@types/natural-compare` as a devDependency to fix a type error introduced by the TypeScript v6 breaking change.

### `tsconfig.json`

- Added `"rootDir": "./src"` to address the breaking change in TypeScript v6.
- Turned off `strictNullChecks` because adopting it would require extensive changes and it was previously disabled.
- Turned off `useUnknownInCatchVariables` because adopting it requires some changes and it was previously disabled.

## Related Issues

N/A

N/A


snitin315 (dismissed):
LGTM, thanks!

src/index.js
```diff
@@ -30,17 +30,17 @@ const plugin = {
...
+			}),
 			rules: recommendedRules,
 		},
 	},
 };
 
-// eslint-disable-next-line no-lone-blocks -- The block syntax { ... } ensures that TypeScript does not get confused about the type of `plugin`.
-{
-	plugin.configs.recommended.plugins.json = plugin;
-}
```
fasttime, line 42: A note that we could also use `Object.assign` here to retain the value property if we don't want to change runtime logic:

```js
Object.assign(plugin.configs.recommended.plugins, { json: plugin });
```

nzakas:
Practical question: TypeScript v6 just came out, do we need to rush to support it? What happens if someone with TypeScript v6 tries to use a v5 package? Is there any end-user benefit or is this just about upgrading to the latest?

src/index.js
lumirlumir: Thanks! The logic looks much cleaner, and the output `.d.ts` file and other logic are the same as before.

I've updated it in https://github.com/eslint/json/pull/224/changes/2d19c3b1e4797670f11e442fd4f74e8e36d64e5f.

lumirlumir:
> Practical question: TypeScript v6 just came out, do we need to rush to support it? What happens if someone with TypeScript v6 tries to use a v5 package? Is there any end-user benefit or is this just about upgrading to the latest?

I first tried upgrading TypeScript to v6 to see whether the deprecation of some options in v6 would affect our use case and end users, since those deprecations might have an impact.

After investigating the codebase and reviewing the [TypeScript v6 announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0), my understanding is that our current `tsconfig.json` setup is not affected by the changes in TypeScript v6. I verified that the generated `.d.ts` files did not change after upgrading to TypeScript v6 and the `.js` files only include refactoring and contain no logic changes. (But, I only verified this against the Markdown, JSON, and CSS language plugins, so other repositories may need to be checked.)

In summary, I think updating to TypeScript v6 will not affect end users because the output files remain unchanged. So, as a result, this PR simply upgrades to the latest version.

fasttime (approved):
The changes LGTM. It seems that the generated types are the same as when building with TypeScript 5.9. Just a note that it would be good to make sure that the type tests still pass with older TypeScript versions (e.g. 5.3), but that could be done in a follow-up PR.

PR #230, opened 2026-04-14
ci: add type tests for TypeScript 5.x to CI workflow

fasttime:
#### Prerequisites checklist

- [X] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

#### AI acknowledgment

- [X] I did _not_ use AI to generate this PR.

add tests in CI

Add type tests for TypeScript v5.3 and v.5.x besides existing tests targeting TypeScript v6.0.

#### Related Issues

refs #224

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,10 @@ jobs:
               run: npm install
             - name: Check Types
               run: npm run test:types
+            - name: Check Types (TypeScript 5.3)
+              run: npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json
+            - name: Check Types (TypeScript 5.x)
+              run: npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json
     are_the_types_wrong:
         name: Are the types wrong?
         runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,9 +64,9 @@ jobs:
             - name: Check Types
               run: npm run test:types
             - name: Check Types (TypeScript 5.3)
-              run: npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json
+              run: npm run test:types:5.3
             - name: Check Types (TypeScript 5.x)
-              run: npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json
+              run: npm run test:types:5.x
     are_the_types_wrong:
         name: Are the types wrong?
         runs-on: ubuntu-latest
--- a/tests/types/tsconfig.dist.json
+++ b/tests/types/tsconfig.dist.json
@@ -1,0 +1,16 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "checkJs": false, // The JSDoc @import tag used in JS files is not supported in TypeScript < 5.5.
+    "noEmit": true,
+    "rootDir": "../..",
+    "strict": true,
+    "strictNullChecks": true,
+    "useUnknownInCatchVariables": true,
+    "exactOptionalPropertyTypes": true,
+    "verbatimModuleSyntax": true
+    // "erasableSyntaxOnly" is not supported in TypeScript < 5.8.
+  },
+  "files": [],
+  "include": ["../../dist"]
+}
--- a/tests/types/tsconfig.dist.json
+++ b/tests/types/tsconfig.dist.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "allowJs": false,
     "checkJs": false, // The JSDoc @import tag used in JS files is not supported in TypeScript < 5.5.
     "noEmit": true,
     "rootDir": "../..",
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -107,8 +107,9 @@ export default defineConfig([
 	{
 		name: "json/jsonc",
 		plugins: { json },
-		files: ["**/*.jsonc"],
+		files: ["**/*.jsonc", "**/tsconfig*.json"],
 		language: "json/jsonc",
+		languageOptions: { allowTrailingCommas: true },
 		extends: ["json/recommended"],
 	},
 	{
--- a/package.json
+++ b/package.json
@@ -57,8 +57,8 @@
     "test:coverage": "c8 npm test",
     "test:jsr": "npx -y jsr@latest publish --dry-run",
     "test:types": "tsc -p tests/types/tsconfig.json",
-    "test:types:5.3": "npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json",
-    "test:types:5.x": "npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json",
+    "test:types:5.3": "npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.legacy.json",
+    "test:types:5.x": "npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.json",
     "test:types:all": "npm run test:types && npm run test:types:5.x && npm run test:types:5.3"
   },
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -56,7 +56,10 @@
     "test": "mocha \"tests/**/*.test.js\"",
     "test:coverage": "c8 npm test",
     "test:jsr": "npx -y jsr@latest publish --dry-run",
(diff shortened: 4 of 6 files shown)


tests/types/tsconfig.dist.json
fasttime, line 1: GitHub marks the comments in this file in red because it doesn't recognize `tsconfig.dist.json` as a TypeScript config file, so it falls back to plain JSON instead of JSONC. The syntax highlighting disappears if the file is renamed to `dist.tsconfig.json`, although that would not align with our conventions.

.github/workflows/ci.yml
```diff
@@ -63,6 +63,10 @@ jobs:
               run: npm install
             - name: Check Types
               run: npm run test:types
+            - name: Check Types (TypeScript 5.3)
+              run: npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json
+            - name: Check Types (TypeScript 5.x)
+              run: npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json
```
lumirlumir, line 69:
```suggestion
            - name: Check Types (TypeScript 5.3)
              run: npm run test:types:5.3
            - name: Check Types (TypeScript 5.x)
              run: npm run test:types:5.x
```

What do you think about moving these scripts to `package.json`’s `scripts` field so it’s possible to test different TypeScript versions locally as well?

For example, if we add `test:types:5.x` and `test:types:5.3` scripts like the following, users could also verify things locally before running them in CI.

https://github.com/eslint/json/blob/ac216ccf95f56ee22b55c94ebefb44a8990d87d8/package.json#L59

```diff
     "test:jsr": "npx -y jsr@latest publish --dry-run",
     "test:types": "tsc -p tests/types/tsconfig.json"
+    "test:types:5.x": "npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json"
+    "test:types:5.3: "npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json"
```

---

Or maybe the following pattern would be worth a look if we want to check type compatibility all in one go:

```diff
     "test:jsr": "npx -y jsr@latest publish --dry-run",
+    "test:types": "npm run test:types:6.x && npm run test:types:5.x && npm run test:types:5.3",
+    "test:types:6.x": "tsc -p tests/types/tsconfig.json"
+    "test:types:5.x": "npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json"
+    "test:types:5.3: "npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json"
```

lumirlumir, line 69:
```suggestion
              run: npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.json
```

Another question: should TypeScript 5.x use `tests/types/tsconfig.dist.json` instead of `tests/types/tsconfig.json`?

Since the latest 5.x version will be TypeScript 5.9.3, and the options disabled in `tests/types/tsconfig.dist.json` appear to be limited to TypeScript 5.3, I think using `tests/types/tsconfig.json` would make the type tests stricter.

tests/types/tsconfig.dist.json
```diff
@@ -0,0 +1,16 @@
...
+    "rootDir": "../..",
+    "strict": true,
+    "strictNullChecks": true,
+    "useUnknownInCatchVariables": true,
+    "exactOptionalPropertyTypes": true,
+    "verbatimModuleSyntax": true
+    // "erasableSyntaxOnly" is not supported in TypeScript < 5.8.
+  },
+  "files": [],
+  "include": ["../../dist"]
```
lumirlumir, line 16:
```suggestion
  "include": ["**/*.test.ts", "../../dist"]
```

Would adding `"**/*.test.ts"` to the `include` pattern make sense here?

Currently, if a type error occurs in `tests/types/types.test.ts`, TS 5.3 silently ignores it. I think adding this pattern would make the type tests more robust.

lumirlumir (changes requested):

tests/types/tsconfig.dist.json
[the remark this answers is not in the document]
lumirlumir: I'm not sure about this approach. I think `checkJS: false` (and `allowJS: false`) do the same thing here.

.github/workflows/ci.yml
fasttime: Good idea. Maybe we could add a `test:types:all` script to run type tests for all TypeScript versions and leave `test:types` as it is? The tests for TypeScript 5.3 and 5.x require fetching packages from the npm registry, which makes them a bit cumbersome for regular use in development, but otherwise they can run locally as well.

fasttime: The main difference is that `tests/types/tsconfig.json` type checks also the JSDoc types in the .js files generated by TypeScript and I'm not sure if there is a reason to do that. It seems that any problems in those types would be already caught either during compilation, or when running the type tests on the corresponding `.d.ts` files.

My preference would be to add `"checkJs": false` to `tests/types/tsconfig.json` as well. Then the only difference between `tests/types/tsconfig.dist.json` and `tests/types/tsconfig.json` would be the `erasableSyntaxOnly` options, which could be used in the type tests for 5.x and 6.x.

tests/types/tsconfig.dist.json
fasttime: Makes sense. There is also a `.cts` file in the same directory, which could be included as well.

.github/workflows/ci.yml
lumirlumir: > Good idea. Maybe we could add a test:types:all script to run type tests for all TypeScript versions and leave test:types as it is?

I understand this as the following approach (or something similar), and it sounds good to me. 👍

```diff
     "test:jsr": "npx -y jsr@latest publish --dry-run",
     "test:types": "tsc -p tests/types/tsconfig.json",
+    "test:types:5.x": "npx -p typescript@5.x -y -- tsc -p tests/types/tsconfig.dist.json"
+    "test:types:5.3": "npx -p typescript@5.3 -y -- tsc -p tests/types/tsconfig.dist.json"
+    "test:types:all": "npm run test:types && npm run test:types:5.x && npm run test:types:5.3"
```

tests/types/tsconfig.dist.json
lumirlumir: > Makes sense. There is also a .cts file in the same directory, which could be included as well.

I think adding a `.cts` file as well would be a good idea 👍

.github/workflows/ci.yml
lumirlumir: Thanks for the clarification. If I understand correctly, would the final `tsconfig.json` and `tsconfig.dist.json` look like the following?

If I understand correctly, I think that would be a good option.

- "TS 6.x" and "TS 5.x" use `tsconfig.json`, and its config would be as follows:

```diff
// tsconfig.json

{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
+   "allowJs": false,
+   "checkJs": false,
    "noEmit": true,
    "rootDir": "../..",
    "strict": true,
    "strictNullChecks": true,
    "useUnknownInCatchVariables": true,
    "exactOptionalPropertyTypes": true,
    "verbatimModuleSyntax": true,
    "erasableSyntaxOnly": true
  },
  "files": [],
  "include": ["**/*.test.ts", "**/*.test.cts", "../../dist"]
}
```

- "TS 5.3" uses `tsconfig.dist.json`, and its config would be as follows:

```diff
// tsconfig.dist.json

{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
+   "allowJs": false,
+   "checkJs": false,
    "noEmit": true,
    "rootDir": "../..",
    "strict": true,
    "strictNullChecks": true,
    "useUnknownInCatchVariables": true,
    "exactOptionalPropertyTypes": true,
    "verbatimModuleSyntax": true
+   // "erasableSyntaxOnly" is not supported in TypeScript < 5.8.
  },
  "files": [],
  "include": ["../../dist"]
}
```

fasttime: Yes, that's basically what I had in mind.

fasttime: Updated in b66b898f26501b05e03ea25d75ba44f712b0e245.

fasttime: Updated in 2ddcba8448f1df3f79a3412339369fe566980fd1. I've also renamed `tsconfig.dist.json` to `tsconfig.legacy.json` to better reflect usage.

tests/types/tsconfig.dist.json
fasttime: Done in 2ddcba8448f1df3f79a3412339369fe566980fd1, thanks for the suggestion.

lumirlumir (approved):
LGTM, thanks!

PR #235, opened 2026-05-05
fix: sort-keys autofix fully sorts keys in a single pass

crimsonjay0:
#### Prerequisites checklist

#### AI acknowledgment

The `sort-keys` autofix previously used a pairwise swap strategy, where each `--fix` pass only swapped two adjacent out-of-order members. For a fully reversed object with N keys, this required up to N-1 passes of `eslint --fix` to reach a fully sorted state.

 I replaced the pairwise swap fixer with a segment-based sorting approach that fully sorts all keys in a single pass.

#### Related Issues

Fixes #222


crimsonjay0:
Not stale. I just forgot this was a draft.

DMartens:
This PR has too many unnecessary unrelated changes to making the fix sort in a single pass.
Can you please limit the changes to the `context.fix` method as there is no need to precompute the fixes if there are none (which should be vast majority of cases)?
You can just use `break` after reporting to only report once.

> Note that @humanwhocodes/momoa doesn't include comments in the object.members tree, so we can't just see if a member is preceded by a comment
You can use `sourceCode.getTokenBefore(property, { includeComents: true })` for this.

src/rules/sort-keys.js
```diff
@@ -250,25 +247,80 @@ const rule = {
...
+									while (
+										end < node.members.length - 1 &&
+										!isLineSeparated(
+											node.members[end],
+											node.members[end + 1],
+										)
+									) {
+										end++;
+									}
+									group = node.members.slice(start, end + 1);
```
DMartens, line 273: Note for other reviewers: This would change the current behavior of reporting every unsorted key but only fixing the first group (because of the `break` after the `context.report`). I think this behavior is okay as the other groups will be fixed in subsequent rounds. The alternative would be to group everything and do the fixes for every group.

src/rules/sort-keys.js
```diff
@@ -250,25 +247,80 @@ const rule = {
...
-									fixer.replaceText(
-										member,
-										sourceCode.getText(prevMemberNode),
-									),
-									fixer.replaceText(
-										prevMemberNode,
-										sourceCode.getText(member),
-									),
-								];
+								const sortedGroup = [...group].sort((a, b) => {
```
DMartens:
```suggestion
								const sortedGroup = group.toSorted((a, b) => {
```
This would be more idiomatic.

src/rules/sort-keys.js
```diff
@@ -250,25 +247,80 @@ const rule = {
...
+									const nameA = getKey(a);
+									const nameB = getKey(b);
+
+									if (nameA === nameB) {
```
DMartens: Can you please add a test case for this?

DMartens:
Thank you for refactoring this, I only have some small notes.

src/rules/sort-keys.js
```diff
@@ -250,25 +247,80 @@ const rule = {
...
 							},
 							fix(fixer) {
-								if (
-									hasAdjacentComment(member) ||
-									hasAdjacentComment(prevMemberNode)
-								) {
-									return null;
+								let group = node.members;
+
+								if (allowLineSeparatedGroups) {
```
DMartens, line 252: Currently there is no test condition which effectively checks that the separated groups before or after are unchanged (`group` is currently always all members)

src/rules/sort-keys.js
```diff
@@ -250,25 +247,80 @@ const rule = {
...
-									fixer.replaceText(
-										prevMemberNode,
-										sourceCode.getText(member),
-									),
-								];
+								const sortedGroup = group.toSorted((a, b) => {
+									const nameA = getKey(a);
+									const nameB = getKey(b);
+
+									if (nameA === nameB) {
```
DMartens: This `if` is unnecessary as the next `if` with the test condition `aFirst && bFirst` also matches with the same return value.

DMartens (approved):
Changes LGTM, thanks. I would like a second review because of the note [above](https://github.com/eslint/json/pull/235#discussion_r3365563810).

DMartens:
Can someone please do a second review @eslint/eslint-team or give their opinion on [whether everything should be fixed at once or a group at a time](https://github.com/eslint/json/pull/235#discussion_r3365563810)?

Tanujkanti4441 (commented):
>This would change the current behavior of reporting every unsorted key but only fixing the first group (because of the break after the context.report). I think this behavior is okay as the other groups will be fixed in subsequent rounds. The alternative would be to group everything and do the fixes for every group.

I feel this change, with only one error message, can be misleading as it is not giving enough information about the unsorted keys specially with the error message where it seems only one key has error:

`Expected object keys to be in alphanumeric case-sensitive ascending order. 'j' should be before 't'.`

I think we can change the error location to the opening curly brace of the object with updated error message something like `Expected object keys to be in alphanumeric case-sensitive ascending order in the root object` for the root object and for nested objects we can change the location to the `key` of the object or from `key` to opening curly brace with message `Expected object keys to be in alphanumeric case-sensitive ascending order in the 'keyName'`?

Also seems the rule is not providing any fix at all if there is any comment between keys with `{ allowLineSeparatedGroups: false }` option and with `{ allowLineSeparatedGroups: true }`,  if there is a second group with a comment then rule doesn't provide any fix for second group so it cannot be fixed in subsequent rounds in this case:

```jsonc
/* eslint json/sort-keys: ["error", "asc", { allowLineSeparatedGroups: true }] */

{
    "t": 1,
    "j": 0,

    "e": 3,
    "d": 5,
    "c": 5,
    //jk
    "n": 3,
    "g": 5,
    "f": 5,
  
    "h": 5,
    "a": 5
}
````
So, we should think about the approach of group everything and do the fixes for every group at the same time.

Also keys with comments between them does not provide any auto-fix as per current behavior and after this change rule cannot even report them with specific error message so users can sort the keys manually.

here we can report the keys individually with current error message?

PR #238, opened 2026-05-10
feat!: add `meta.languages` to JSON rules

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR adds `meta.languages` to rules, following the decision made during the TSC meeting on 16-Apr-2026.

Added `meta.languages` to rule definitions and `meta.docs.dialects` to their docs metadata.

#### Related Issues

https://github.com/eslint/eslint/pull/20716#issuecomment-4355708101


lumirlumir (changes requested):
After further investigation, I've found that this change could lead to a breaking change for end-users in a way different from what was mentioned in [Markdown](https://github.com/eslint/markdown/issues/339#issuecomment-4241381591).

For instance, if we use the default setup described in the `README.md` as shown below, it works correctly.

```js
// eslint.config.js
import { defineConfig } from "eslint/config";

export default defineConfig([
	{
		plugins: {
			json,
		},
	},

	// lint JSON files
	{
		files: ["**/*.json"],
		language: "json/json",
		rules: {
			"json/no-duplicate-keys": "error",
		},
	},

	// lint JSONC files
	{
		files: ["**/*.jsonc", ".vscode/*.json"],
		language: "json/jsonc",
		rules: {
			"json/no-duplicate-keys": "error",
		},
	},

	// lint JSON5 files
	{
		files: ["**/*.json5"],
		language: "json/json5",
		rules: {
			"json/no-duplicate-keys": "error",
		},
	},
]);
```

However, what if a user prefers a different namespace, such as `eslintjson`, instead of `json`?

Suppose we have a configuration like the one below, which uses the `eslintjson` namespace:

```js
// eslint.config.js
import { defineConfig } from "eslint/config";
import eslintjson from "@eslint/json";

export default defineConfig([
	{
		plugins: {
			eslintjson,
		},
	},

	// lint JSON files
	{
		files: ["**/*.json"],
		language: "eslintjson/json",
		rules: {
			"eslintjson/no-duplicate-keys": "error",
		},
	},

	// lint JSONC files
	{
		files: ["**/*.jsonc", ".vscode/*.json"],
		language: "eslintjson/jsonc",
		rules: {
			"eslintjson/no-duplicate-keys": "error",
		},
	},

	// lint JSON5 files
	{
		files: ["**/*.json5"],
		language: "eslintjson/json5",
		rules: {
			"eslintjson/no-duplicate-keys": "error",
		},
	},
]);
```

The newly added `languages: ["json/json", "json/jsonc", "json/json5"]` field prevents the above configuration from working and throws an error like this:

```txt
Oops! Something went wrong! :(

ESLint: 10.3.0

The following rules do not support the language "eslintjson/json":
        - "eslintjson/no-empty-keys"

To fix this error, either:
- Remove the rule from your configuration, or set its severity to "off".
- Use the "files" option to apply the rule only to files of the supported language, for example:
  {
    files: ["**/*.js"],
    rules: { "eslintjson/no-empty-keys": "error" }
  }

See https://eslint.org/docs/latest/use/configure/rules for more information.
```

Before adding the `languages` field, this configuration worked correctly. It only fails after the addition of the `languages` field.

---

It seems this implies that the newly added `languages` field restricts the namespace to only those explicitly described. Previously, users were not restricted by the namespace used in the rule, allowing them to use `eslintjson`, `eslintcorejson`, or any other prefix they preferred instead of `json`.

For example, if a user uses two different JSON plugins, such as [`eslint-plugin-json`](https://github.com/azeemba/eslint-plugin-json), they might want to use distinct namespaces for each—like `json1/rule-name` for `@eslint/json` and `json2/rule-name` for `eslint-plugin-json`.

This change could potentially break that scenario.

DMartens:
The problem lumir mentioned could be solved if the core normalizes the language id by using `plugin.meta.name` (here "@eslint/json") instead of the plugin id.
This normalization would be similar to the plugin name inference in eslintrc's `plugin` (e.g. for scoped packages only the package name, which would be "json").

nzakas:
Yes, this is a bug in the core implementation. It should normalize the namespace used for language resolution. Can you open a bug for that?

lumirlumir:
It seems like an issue hasn't been opened yet. I can open one.

fasttime (commented):
We need to add `namespace: "json"` in the plugin's `meta` object to avoid the problem described in https://github.com/eslint/json/pull/238#pullrequestreview-4262066397.

https://github.com/eslint/json/blob/720c43d6876ca5eda969f4b6c3882fcdcd960a52/src/index.js#L20-L23

lumirlumir (changes requested):
I think it would also be helpful to add an integration test in `tests/plugin/eslint.test.js` to verify the behavior mentioned in https://github.com/eslint/json/pull/238#pullrequestreview-4262066397 and prevent any potential regressions.

https://github.com/eslint/json/blob/main/tests/plugin/eslint.test.js

package.json
```diff
@@ -82,7 +82,7 @@
     "@types/node": "^20.19.0",
     "c8": "^11.0.0",
     "dedent": "^1.5.3",
-    "eslint": "^10.0.0",
+    "eslint": "^10.4.0",
```
Pixel998, line 85: Updated ESLint so the test runs with `meta.languages` enforcement, which was added in 10.2.0.

fasttime (commented):
I think this PR should be considered a breaking change for this plugin and tagged as `feat!`, because it may break compatibility with existing configs, even if that's a desired effect. Other than that, the changes LGTM.

lumirlumir (approved):
LGTM, thanks! 

> I think this PR should be considered a breaking change for this plugin and tagged as feat!, because it may break compatibility with existing configs, even if that's a desired effect.

While I can no longer find a certain failure scenario with this setup, there might be some edge cases we’re all missing. I think it may make sense to defer this change to the next major version for safety.

fasttime (approved):
LGTM, thanks! I think we don't need to wait for a major ESLint release to merge this PR because `@eslint/json` isn't a production dependency of ESLint.

Leaving open for another approval from a TSC member since this is a breaking change.

lumirlumir:
>  I think we don't need to wait for a major ESLint release to merge this PR because @eslint/json isn't a production dependency of ESLint.

I should have been clearer. What I meant was the major release of `@eslint/json`, which is v2.

---

I have one quick question: Are major releases of `@eslint/json` and `@eslint/css` planned anytime soon? I have an issue in mind to discuss for the next major release of these language plugins, so I wanted to ask whether that is currently planned.

nzakas:
@lumirlumir we don't plan major releases of plugins in advance, we do them when we need them. Open any issue you want and we can discuss.

nzakas (approved):
LGTM. Thanks.

PR #248, opened 2026-06-19
fix: type compatibility issue between JSON v2 and ESLint v9

### LINKED ISSUE #213: Bug: JSON v1.0.1 is not backward-compatible with ESLint v9
### Environment

ESLint version: 9.39.2
@eslint/json version: 1.0.1
Node version: 20.19.1
npm version: 10.8.2
Operating System: Stackblitz

json

### What did you do?

```js
// eslint.config.js

import json from '@eslint/json';
import { defineConfig } from 'eslint/config';

export default defineConfig([
  {
    files: ['**/*.jsonc'],
    plugins: { json },
    language: 'json/jsonc',
    extends: ['json/recommended'],
  },
]);
```

```jsonc
// tsconfig.json

{
  "include": ["eslint.config.js"],
  "compilerOptions": {
    /* Type Checking */
    "allowUnreachableCode": false,
    "allowUnusedLabels": false,
    "alwaysStrict": true,
    "exactOptionalPropertyTypes": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitAny": true,
    "noImplicitOverride": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noPropertyAccessFromIndexSignature": false, 
    "noUncheckedIndexedAccess": false,
    "noUnusedLocals": false,
    "noUnusedParameters": true,
    "strict": true,
    "strictBindCallApply": true,
    "strictBuiltinIteratorReturn": true,
    "strictFunctionTypes": true,
    "strictNullChecks": true,
    "strictPropertyInitialization": true,
    "useUnknownInCatchVariables": true,

    /* Modules */
    "module": "NodeNext",
    "moduleResolution": "nodenext",

    /* JavaScript Support */
    "allowJs": true,
    "checkJs": true,

    /* Language and Environment */
    "target": "ESNext"
  }
}
```

I expected the types in `@eslint/json@1.0.1` to be compatible with ESLint v9.

According to the CI test strategy, `README.md`, and the comments in markdown (https://github.com/eslint/markdown/pull/593#discussion_r2784855613), JSON v1 is expected to remain backward compatible with ESLint v9.

https://github.com/eslint/json/blob/4106331b746c7944b869a9b663b0d1c4838986df/.github/workflows/ci.yml#L30

### What actually happened?

The types in JSON v1 are not compatible with ESLint v9, which prevents using this plugin with ESLint v9 despite it being mentioned as compatible.

Image

https://stackblitz.com/edit/stackblitz-starters-kpy9aaea?file=eslint.config.js

Please run `npm run tsc` at the reproduction link.

### Participation

### Additional comments

The problem is that the `ESLint.plugin` type, which is extracted from `import type { ESLint } from "eslint";`, now has different type signatures:

- ESLint v9.39.2 depends on `@eslint/core` 0.17.0
- ESLint v10.0.0 depends on `@eslint/core` 1.1.0
- JSON v0.14.0 depends on `@eslint/core` 0.17.0
- JSON v1.0.1 depends on `@eslint/core` 1.1.0

So, 

- JSON v1 and ESLint v10 is compatible. ✅ 
- JSON v1 and ESLint v9 is incompatible ❌ 

---

Another point is that the HEAD branches of CSS and Markdown are affected by this problem. I discovered it while working on https://github.com/eslint/markdown/pull/593.

To summarize, my question is: do we need to retain type compatibility with ESLint v9 for JSON v1 / CSS v1 / and Markdown v8? If so, I think we need a fundamental solution to ensure these types work seamlessly. Since one of the primary aims of the major releases of language plugins was to maintain backward compatibility, I hope this issue can be addressed.

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

## What is the purpose of this pull request?

This PR fixes an issue where JSON v2 and ESLint v9 are not type-compatible, as mentioned in eslint/json#213.

This PR follows the same strategy determined in https://github.com/eslint/markdown/pull/648. That said, a few parts differ from the Markdown implementation, as follows:

1. For JSON, the type compatibility range is 9.15.0 or above, not 9.39.0. This is because the [Markdown plugin has a `processor`](https://github.com/eslint/markdown/blob/main/src/index.js#L100), and the blocker for compatibility with lower ESLint v9 versions was that `processor`. Since CSS does not have a `processor` configured, broader type compatibility can be achieved.
1. As the type compatibility range is now 9.15.0 or above, I intentionally didn’t mention in `README.md` that type compatibility is ensured for v9.15.0 or above, since the current type compatibility range is the same as the one we already support.

## Related Issues

Refs: https://github.com/eslint/markdown/pull/648, https://github.com/eslint/css/pull/473
Fixes: #213

N/A


src/rules/no-duplicate-keys.js
```diff
@@ -49,7 +48,7 @@ const rule = {
 		/** @type {Map<string, MemberNode>|undefined} */
 		let keys;
 
-		return {
+		return /** @type {JSONRuleVisitor} */ ({
```
lumirlumir, line 51: Unlike the CSS implementation, if we don’t type-cast the return type with `JSONRuleVisitor`, a portable type error occurs:

```txt
The inferred type of 'create' cannot be named without a reference to 'MemberNode' from '../../node_modules/@humanwhocodes/momoa/dist/typedefs.js'. This is likely not portable. A type annotation is necessary.
```

This was also why only the JSON plugin used the following snippet:

```js
const rule = { /* ... */ };

export default rule;
```

Instead of:

```js
export default { /* ... */ };
```

---

I’ve tried to find an easy way to resolve it, but I couldn’t find one we could handle on our own.

I think this issue lies with `@humanwhocodes/momoa`, since the package still uses a dual CJS/ESM setup and type resolution seems to have broken at some point.

Sending a PR to `@humanwhocodes/momoa` to make it ESM-only may resolve the type issue, since the portable error seems to come from [`typedefs.ts`](https://npmx.dev/package-code/@humanwhocodes/momoa/v/3.3.10/dist%2Ftypedefs.ts) not being exported directly from the main entry point `./dist/momoa.js`, but instead being exported indirectly.

fasttime: I'm not sure how to export `dist/typedefs.ts` directly from `dist/momoa.js`. Do you have an idea?

lumirlumir: What I found is that in the distributed type declaration file [`dist/momoa.d.ts`](https://npmx.dev/package-code/@humanwhocodes/momoa/v/3.3.10/dist%2Fmomoa.d.ts) from `@humanwhocodes/momoa`, if we remove `export type AnyNode = import("./typedefs.js").AnyNode;` and instead explicitly import and export the types (for example, `import type { AnyNode } from "./typedefs.js";` and `export type { AnyNode } from "./typedefs.js";`), the portable type error is resolved.

Here is a minimal patch I applied directly in `node_modules`, and it worked as expected:

<details>
<summary>Patched File</summary>

```ts
import type {
    AnyNode,
    JSONValue,
    Token,
    TokenizeOptions,
    NodeParts,
    DocumentNode,
    StringNode,
    NumberNode,
    BooleanNode,
    MemberNode,
    ObjectNode,
    ElementNode,
    ArrayNode,
    NullNode,
    ValueNode,
    IdentifierNode,
    NaNNode,
    InfinityNode,
    Sign,
    Node,
    ParseOptions,
    TraversalPhase,
} from "./typedefs.js";
export type {
    AnyNode,
    JSONValue,
    TokenType,
    Location,
    Token,
    Range,
    TokenizeOptions,
    LocationRange,
    NodeParts,
    DocumentNode,
    StringNode,
    NumberNode,
    BooleanNode,
    MemberNode,
    ObjectNode,
    ElementNode,
    ArrayNode,
    NullNode,
    ValueNode,
    IdentifierNode,
    NaNNode,
    InfinityNode,
    Sign,
    Node,
    Mode,
    ParseOptions,
    TraversalPhase,
} from "./typedefs.js";
export type FilterPredicate = (item: {
    node: Node;
    parent?: Node;
    phase: TraversalPhase;
}, index: number, array: Array<{
    node: Node;
    parent?: Node;
    phase: TraversalPhase;
}>) => boolean;
export type TraversalVisitor = {
    enter?: (node: Node, parent?: Node) => void;
    exit?: (node: Node, parent?: Node) => void;
};

/**
 * @fileoverview Evaluator for Momoa AST.
 * @author Nicholas C. Zakas
 */
/** @typedef {import("./typedefs.js").AnyNode} AnyNode */
/** @typedef {import("./typedefs.js").JSONValue} JSONValue */
/**
 * Evaluates a Momoa AST node into a JavaScript value.
 * @param {AnyNode} node The node to interpet.
 * @returns {JSONValue} The JavaScript value for the node.
 */
export function evaluate(node: AnyNode): JSONValue;
/**
 * @callback FilterPredicate
 * @param {{node: Node, parent?: Node, phase: TraversalPhase}} item
 * @param {number} index
 * @param {Array<{node: Node, parent?: Node, phase: TraversalPhase}>} array
 * @returns {boolean}
 */
/**
 * Creates an iterator over the given AST.
 * @param {Node} root The root AST node to traverse.
 * @param {FilterPredicate} [filter] A filter function to determine which steps to
 *      return;
 * @returns {IterableIterator<{node: Node, parent?: Node, phase: TraversalPhase}>} An iterator over the AST.
 */
export function iterator(root: Node, filter?: FilterPredicate): IterableIterator<{
    node: Node;
    parent?: Node;
    phase: TraversalPhase;
}>;
/**
 *
 * @param {string} text The text to parse.
 * @param {ParseOptions} [options] The options object.
 * @returns {DocumentNode} The AST representing the parsed JSON.
 * @throws {Error} When there is a parsing error.
 */
export function parse(text: string, options?: ParseOptions): DocumentNode;
/**
 * Converts a Momoa AST back into a JSON string.
 * @param {AnyNode} node The node to print.
 * @param {Object} options Options for the print.
 * @param {number} [options.indent=0] The number of spaces to indent each line. If
 *      greater than 0, then newlines and indents will be added to output.
 * @returns {string} The JSON representation of the AST.
 */
export function print(node: AnyNode, { indent }?: {
    indent?: number;
}): string;
/**
 * Creates an iterator over the tokens representing the source text.
 * @param {string} text The source text to tokenize.
 * @param {TokenizeOptions} [options] Options for doing the tokenization.
 * @returns {Array<Token>} An iterator over the tokens.
 */
export function tokenize(text: string, options?: TokenizeOptions): Array<Token>;
/**
 * Traverses an AST from the given node.
 * @param {Node} root The node to traverse from
 * @param {TraversalVisitor} visitor An object with an `enter` and `exit` method.
 */
export function traverse(root: Node, visitor: TraversalVisitor): void;
export namespace types {
    /**
     * Creates a document node.
     * @param {ValueNode} body The body of the document.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {DocumentNode} The document node.
     */
    export function document(body: ValueNode, parts?: NodeParts): DocumentNode;
    /**
     * Creates a string node.
     * @param {string} value The value for the string.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {StringNode} The string node.
     */
    export function string(value: string, parts?: NodeParts): StringNode;
    /**
     * Creates a number node.
     * @param {number} value The value for the number.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {NumberNode} The number node.
     */
    export function number(value: number, parts?: NodeParts): NumberNode;
    /**
     * Creates a boolean node.
     * @param {boolean} value The value for the boolean.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {BooleanNode} The boolean node.
     */
    export function boolean(value: boolean, parts?: NodeParts): BooleanNode;
    /**
     * Creates a null node.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {NullNode} The null node.
     */
    function _null(parts?: NodeParts): NullNode;
    export { _null as null };
    /**
     * Creates an array node.
     * @param {Array<ElementNode>} elements The elements to add.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {ArrayNode} The array node.
     */
    export function array(elements: Array<ElementNode>, parts?: NodeParts): ArrayNode;
    /**
     * Creates an element node.
     * @param {ValueNode} value The value for the element.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {ElementNode} The element node.
     */
    export function element(value: ValueNode, parts?: NodeParts): ElementNode;
    /**
     * Creates an object node.
     * @param {Array<MemberNode>} members The members to add.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {ObjectNode} The object node.
     */
    export function object(members: Array<MemberNode>, parts?: NodeParts): ObjectNode;
    /**
     * Creates a member node.
     * @param {StringNode|IdentifierNode} name The name for the member.
     * @param {ValueNode} value The value for the member.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {MemberNode} The member node.
     */
    export function member(name: StringNode | IdentifierNode, value: ValueNode, parts?: NodeParts): MemberNode;
    /**
     * Creates an identifier node.
     * @param {string} name The name for the identifier.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {IdentifierNode} The identifier node.
     */
    export function identifier(name: string, parts?: NodeParts): IdentifierNode;
    /**
     * Creates a NaN node.
     * @param {Sign} sign The sign for the Infinity.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {NaNNode} The NaN node.
     */
    export function nan(sign?: Sign, parts?: NodeParts): NaNNode;
    /**
     * Creates an Infinity node.
     * @param {Sign} sign The sign for the Infinity.
     * @param {NodeParts} parts Additional properties for the node.
     * @returns {InfinityNode} The Infinity node.
     */
    export function infinity(sign?: Sign, parts?: NodeParts): InfinityNode;
}
/**
 * @fileoverview Traversal approaches for Momoa JSON AST.
 * @author Nicholas C. Zakas
 */
/** @typedef {import("./typedefs.js").TraversalPhase} TraversalPhase */
/**
 * @typedef {Object} TraversalVisitor
 * @property {(node: Node, parent?: Node) => void} [enter]
 * @property {(node: Node, parent?: Node) => void} [exit]
 */
declare const childKeys: Map<string, string[]>;
export { childKeys as visitorKeys };
```

</details>

image

---

But the problem is that, in the current state of `@humanwhocodes/momoa`, the build process depends on [`rollup`](https://github.com/humanwhocodes/momoa/blob/main/js/rollup.config.js), similar to how we do it in the `rewrite` repository, so there doesn’t seem to be an easy way to resolve this issue.

One option would be to jump through multiple hoops to produce the type declaration file we want when building declarations in `@humanwhocodes/momoa`. But based on https://github.com/eslint/rewrite/issues/403#issuecomment-4060155513, I’m not sure that would be the preferred approach.

Another option would be to bump `@humanwhocodes/momoa` to v4 and make it ESM-only with an updated Node.js support range and apply the same feature from https://github.com/eslint/rewrite/issues/403.

I think resolving this issue clearly will require some effort. But as a practical option for now, I believe we could track the `JSONRuleVisitor` type cast in a separate issue and remove it once this is resolved in `@humanwhocodes/momoa`. What do you think about this direction?

lumirlumir:
Friendly ping @fasttime

src/rules/no-duplicate-keys.js
fasttime: I see. So using type export statements instead of `import()`-based type references avoids the compiler error when TypeScript can't infer a portable return type. That still feels a bit surprising, because the `import()` syntax doesn't seem problematic in the other cases, and it also appears to work fine when the referenced module isn't under `node_modules`.

I don't think we need to track the `JSONRuleVisitor` cast separately, since ideally we'd remove this workaround altogether at some point. The type cast should be good enough until then.

As for updating the `@humanwhocodes/momoa` types, I'm not sure either. My current understanding is that this is related to how the types are declared in JSDoc (with `/** @typedef {import("./typedefs.js").MemberNode} MemberNode */`), rather than something specific to CommonJS transpilation or Rollup. Maybe you could open a separate issue to discuss the change you have in mind in `@humanwhocodes/momoa` sepecifically.

lumirlumir: Thanks, I've understood the situation. I opened a separate issue in `@humanwhocodes/momoa` to discuss what I have in mind: https://github.com/humanwhocodes/momoa/issues/236.

Is there anything else in particular that I need to update?

fasttime (approved):
LGTM, thanks!

PR #194, opened 2025-12-21
fix: program crashes in `no-unsafe-values`

lumirlumir:
## Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

### Which language are you using?

JSON5

### What did you do?

I expected the following cases not to crash the program, but they do in JSON5 mode:

```json5
0x0
```

```json5
.0
```

```json5
+1e-400
```

(... and additional cases listed in `tests/rules/no-unsafe-values.test.js`.)

Also, I expected the following cases to be reported as `unsafeZero`, but they aren't:

```json5
+1e-400
```

```json5
.01e-400
```

```json5
-10.e-402
```

(... and additional cases listed in `tests/rules/no-unsafe-values.test.js`.)

### What did you expect to happen?

I expect the above cases not to crash the program and some of them to be reported as `unsafeZero`.

### Link to minimal reproducible Example

The following code would help identify the problem:

- Crash caused by a valid `+` token in JSON5 mode. The crash is triggered by `+1e-400` (Also, `+1e-400` should be reported as an error once the crash is resolved.)

```json5
// eslint json/no-unsafe-values: "error"

[1e-400, +1e-400, -1e-400]
```

- Crash caused by leading decimal point in JSON5 mode. (Also, `.01e-400` should be reported as an error once the crash is resolved.)

```json5
// eslint json/no-unsafe-values: "error"

.01e-400
```

- Crash caused by a valid `0x`/`0X` hexadecimal prefix in JSON5 mode.

```json5
// eslint json/no-unsafe-values: "error"

0x0
```

## What changes did you make? (Give an overview)

According to the JSON5 specification, JSON5 supports a wider range of numeric tokens or prefixes than JSON and JSONC. 

https://spec.json5.org/#summary-of-features

image

For example, the following numeric are valid:

(The following example doesn't work in JSON or JSONC mode; it only works in JSON5 mode.)

- The `+` token for positive numbers:

[...]

PR #256, opened 2026-07-09
refactor: safe fallback for regex groups in no-unsafe-values

### LINKED ISSUE #255: Change request: no-unsafe-values rule should not pass undefined to RegExp.test() for empty fractional capture groups
### Environment

ESLint version: v10.4.0 (or whichever version is active)
@eslint/json version: v2.0.1 (main branch)
Node version: v20.19.0
npm version: v10.8.2
Operating System: Windows 11

json

### What did you do?

<details>

```js

export default [
    {
        plugins: {
            json,
        },
        language: "json/json5",
        rules: {
            "json/no-unsafe-values": "error",
        },
    }
];
```
</details>

I expected the rule to safely evaluate the number without passing `undefined` to `RegExp.prototype.test()`.

In `src/rules/no-unsafe-values.js`, the `NUMBER` regular expression captures the `int` and `frac` portions of a numeric value to determine if an unsafe zero was actually intended to be zero. When the parsed number lacks an integer part (e.g., `.1e-500` in JSON5), the `int` capture group becomes `undefined`.

The rule then executes `NON_ZERO.test(match.groups.int)`, which resolves to `/[1-9]/.test(undefined)`. JavaScript coerces `undefined` into the string `"undefined"`. Because the string `"undefined"` doesn't contain digits 1-9, it evaluates to `false` and coincidentally avoids crashing. However, relying on this implicit type coercion quirk is a fragile code smell that could cause logic bugs if the regex criteria change. 

I expect the groups to be coalesced to an empty string safely like so:
```javascript
if (
    NON_ZERO.test(match.groups.int || "") ||
    NON_ZERO.test(match.groups.frac || "")
)
```

### What actually happened?

The rule internally executes `RegExp.test(undefined)`, causing implicit coercion to `"undefined"`. No runtime crash occurs, but the code relies on fragile, unintended behavior.

https://github.com/eslint/json/blob/main/src/rules/no-unsafe-values.js#L89

### Participation

### AI acknowledgment

### Additional comments

If this issue is accepted, I already have a minimal fix tested locally and I would be happy to open a pull request!

jibin7jose:
Fixes #255

This PR safely coalesces the `int` and `frac` capture groups to an empty string (`""`) inside the `no-unsafe-values` rule before testing them with the `NON_ZERO` regex. 

This prevents JavaScript from implicitly coercing `undefined` into the string `"undefined"` when `RegExp.test()` is called with missing capture groups (e.g., when processing numbers like `.e-100` that omit the integer part).

### DIFF
--- a/src/rules/no-unsafe-values.js
+++ b/src/rules/no-unsafe-values.js
@@ -86,8 +86,8 @@ const rule = {
 						// non-zero digit in it, this number was not intended to be
 						// evaluated down to a zero.
 						if (
-							NON_ZERO.test(match.groups.int) ||
-							NON_ZERO.test(match.groups.frac)
+							NON_ZERO.test(match.groups.int ?? "") ||
+							NON_ZERO.test(match.groups.frac ?? "")
 						) {
 							context.report({
 								loc: node.loc,


src/rules/no-unsafe-values.js
```diff
@@ -86,8 +86,8 @@ const rule = {
 						// non-zero digit in it, this number was not intended to be
 						// evaluated down to a zero.
 						if (
-							NON_ZERO.test(match.groups.int) ||
-							NON_ZERO.test(match.groups.frac)
+							NON_ZERO.test(match.groups.int || "") ||
```
DMartens: These should be nullish coalescing (`match.groups.int ?? ""`) as the group is potentially `undefined` and cannot be another falsy value

DMartens:
In the future please wait for an issue to be accepted before providing a PR and please use the existing PR template.
Can you please sign the EasyCLA (see comment above) as otherwise we cannot move this PR forward.

src/rules/no-unsafe-values.js
jibin7jose: Updated to use nullish coalescing (`??`) as suggested — you're right that the capture groups can only be `undefined`, so `??` is more semantically precise. Thanks for the review!

jibin7jose:
Noted, and apologies for the premature PR — I'll make sure to wait for issue acceptance before submitting PRs in the future. I've also taken note of the PR template requirement and will follow it going forward. Thanks for the guidance!

lumirlumir (changes requested):
I’ve read through the issue, but I’m not convinced that this change provides a concrete benefit.

While working on #194, I also considered that `match.groups.int` and `match.groups.frac` could be `undefined`, but intentionally left the code unchanged.

The relevant cases are already covered by the tests added in that PR, as noted in https://github.com/eslint/json/issues/255#issuecomment-4924511065. Adding nullish coalescing would not change the behavior and, in my view, would make the condition slightly more verbose and less clear.

For these reasons, I’m personally -1 on this change unless there is a concrete behavioral or maintainability benefit that I may have overlooked.

(And I’m marking the PR prefix as `refactor` since this is not a bug fix.)

lumirlumir:
@jibin7jose Could you also disclose your AI usage in this PR, per our [AI Usage Policy](https://eslint.org/docs/latest/contribute/ai-policy)?

jibin7jose:
Hi @lumirlumir, thanks for taking a look and for the context on #194!

I understand your point that the tests already cover these cases and the change doesn't provide a concrete behavioral benefit, making it unnecessarily verbose. Since you're -1 on this, I'll go ahead and close the PR.

Regarding the AI usage: Yes, I used an AI coding assistant to help identify the potential undefined values and to draft this code. Thanks for pointing out the AI Usage Policy, I'll make sure to disclose this properly in any future contributions!

PR #258, opened 2026-07-13
docs: add eslint to install command

minseonkkim:
#### Prerequisites checklist

#### AI acknowledgment

Documentation update.

`@eslint/json` does not declare `eslint` as a peer dependency, so installing the plugin alone does not install ESLint. However, the install commands in the README only include `@eslint/json`, which may lead new users to miss the required `eslint` dependency. This PR updates the install commands to include `eslint` so users can get started without missing dependencies.

Added `eslint` to the install commands in the README for npm, yarn, pnpm, and bun:

- `npm install @eslint/json -D` → `npm install eslint @eslint/json -D`
- `yarn add @eslint/json -D` → `yarn add eslint @eslint/json -D`
- `pnpm install @eslint/json -D` → `pnpm install eslint @eslint/json -D`
- `bun add @eslint/json -D` → `bun add eslint @eslint/json -D`

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@ This package contains a plugin that allows you to natively lint JSON, JSONC, and
 For Node.js and compatible runtimes:
 
 ```shell
-npm install @eslint/json -D
+npm install eslint @eslint/json -D
 # or
-yarn add @eslint/json -D
+yarn add eslint @eslint/json -D
 # or
-pnpm install @eslint/json -D
+pnpm install eslint @eslint/json -D
 # or
-bun add @eslint/json -D
+bun add eslint @eslint/json -D
 ```
 
 For Deno:
--- a/README.md
+++ b/README.md
@@ -8,16 +8,27 @@ This package contains a plugin that allows you to natively lint JSON, JSONC, and
 
 ## Installation
 
-For Node.js and compatible runtimes:
+You can install ESLint using npm or other package managers:
+```shell
+npm install eslint -D
+# or
+yarn add eslint -D
+# or
+pnpm install eslint -D
+# or
+bun add eslint -D
+```
+
+Then, for Node.js and compatible runtimes:
 
 ```shell
-npm install eslint @eslint/json -D
+npm install @eslint/json -D
 # or
-yarn add eslint @eslint/json -D
+yarn add @eslint/json -D
 # or
-pnpm install eslint @eslint/json -D
+pnpm install @eslint/json -D
 # or
-bun add eslint @eslint/json -D
+bun add @eslint/json -D
 ```
 
 For Deno:
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ This package contains a plugin that allows you to natively lint JSON, JSONC, and
 ## Installation
 
 You can install ESLint using npm or other package managers:
+
 ```shell
 npm install eslint -D
 # or


DMartens:
Thank you for filing a PR for this.
I think the proper fix would be to declare the peer dependency on `eslint` rather than updating the documentation as we recommend this for plugins [in the docs](https://eslint.org/docs/latest/extend/plugins#share-plugins).
I could not find issues regarding this but all three official language plugins (`css`, `markdown` and `json`) do not declare the peer dependency.
I would like to know how other team members feel about this @eslint/eslint-team.

fasttime:
> I think the proper fix would be to declare the peer dependency on `eslint` rather than updating the documentation as we recommend this for plugins [in the docs](https://eslint.org/docs/latest/extend/plugins#share-plugins). I could not find issues regarding this but all three official language plugins (`css`, `markdown` and `json`) do not declare the peer dependency. I would like to know how other team members feel about this @eslint/eslint-team.

`@eslint/js` already includes ESLint v10 as an optional peer dependency (https://github.com/eslint/eslint/blob/main/packages/js/package.json#L33-L40):

```json
"peerDependencies": {
  "eslint": "^10.0.0"
},
"peerDependenciesMeta": {
  "eslint": {
    "optional": true
  }
}
```

We could add the same settings for this plugin (and also `@eslint/css` and `@eslint/markdown`) and set an appropriate ESLint version range, for example `"eslint": "^9.15.0 || ^10.0.0"`.

That would be a breaking change, though, because it could break `npm install` for a package that depends on this plugin and also depends on an older version of eslint. A non-breaking alternative would be to set `"eslint": "*"`.

lumirlumir:
The rationale for removing `eslint` from `peerDependencies` -- and for not retaining it as an optional peer dependency through `peerDependenciesMeta` -- comes from https://github.com/eslint/json/pull/25, https://github.com/eslint/markdown/pull/284, and https://github.com/eslint/css/pull/273#issuecomment-3324091645

The [initial commit](https://github.com/eslint/json/pull/25/changes/08fd5d3c41f015689561af762dbc6eec66592add) did include `peerDependenciesMeta` with `eslint` marked as optional, but it was removed following the discussion. As explained in the thread, the primary concern was compatibility with Code Explorer (and perhaps broader JavaScript runtime compatibility)

---

Given the original rationale, I believe we can simply add instructions to install `eslint` at the beginning of the `@eslint/json` installation section. [`@eslint/js` follows the same approach](https://github.com/eslint/eslint/blob/main/packages/js/README.md?plain=1#L30-L40), recommending that users install `eslint` before installing `@eslint/js`. I believe `@eslint/json` should do the same.

---

As for why `@eslint/js` still declares `eslint` as an optional peer dependency: I believe Code Explorer [uses `@eslint/json`, `@eslint/css`, and `@eslint/markdown` in the browser](https://github.com/eslint/code-explorer/blob/main/src/hooks/use-ast.ts#L5-L7), whereas [`@eslint/js` is used only in the Node.js environment through `eslint.config.js`](https://github.com/eslint/code-explorer/blob/main/eslint.config.js#L2).

I think this distinction explains why `@eslint/js` retains `peerDependenciesMeta` with `eslint` marked as optional instead of removing the peer dependency entirely.

---

So, to clarify, I’m +1 on this change. It would also bring the `@eslint/json` installation instructions more closely in line with those for `@eslint/js`.

README.md
```diff
@@ -11,13 +11,13 @@ This package contains a plugin that allows you to natively lint JSON, JSONC, and
 For Node.js and compatible runtimes:
 
 ```shell
-npm install @eslint/json -D
+npm install eslint @eslint/json -D
```
lumirlumir:
```suggestion
npm install @eslint/json -D
```

````diff
+ You can install ESLint using npm or other package managers:
+ 
+ ```shell
+ npm install eslint -D
+ # or
+ yarn add eslint -D
+ # or
+ pnpm install eslint -D
+ # or
+ bun add eslint -D
+ ```
+
+ Then, for Node.js and compatible runtimes:

```shell
npm install @eslint/json -D
# or
yarn add @eslint/json -D
# or
pnpm install @eslint/json -D
# or
bun add @eslint/json -D
```
````

I think that, to align with our existing convention in [`@eslint/js`](https://github.com/eslint/eslint/blob/main/packages/js/README.md?plain=1#L30-L40), adding a dedicated section above this one would be nicer than prefixing `@eslint/json` with `eslint`. Could you update it when you’re available?

**Review by lumirlumir — CHANGES_REQUESTED** [review]

File: README.md (hunk 1 above)
Comment (minseonkkim) [reply README.md]: I've updated it to add a separate ESLint installation section above, following the @eslint/js convention. Thanks for the review!

**Review by lumirlumir — DISMISSED** [review]
LGTM, thanks!

---

It looks good to me, but I’d like to know what @DMartens and @fasttime think about the comment in https://github.com/eslint/json/pull/258#issuecomment-4975796537 before merging.

I believe that’s the original thread for the discussion around `peerDependencies`, but I just wanted to double-check.

File: README.md [hunk 2]
```diff
@@ -8,7 +8,18 @@ This package contains a plugin that allows you to natively lint JSON, JSONC, and
 
 ## Installation
 
-For Node.js and compatible runtimes:
+You can install ESLint using npm or other package managers:
```
lumirlumir, line 11:
```suggestion
You can install ESLint using npm or other package managers:

```

@minseonkkim Could you add a blank line between the paragraph and the code block, or run `npm run fmt` once? The CI failure is caused by this.

lumirlumir (approved):

fasttime:
> As for why `@eslint/js` still declares `eslint` as an optional peer dependency: I believe Code Explorer [uses `@eslint/json`, `@eslint/css`, and `@eslint/markdown` in the browser](https://github.com/eslint/code-explorer/blob/main/src/hooks/use-ast.ts#L5-L7), whereas [`@eslint/js` is used only in the Node.js environment through `eslint.config.js`](https://github.com/eslint/code-explorer/blob/main/eslint.config.js#L2).
> 
> I think this distinction explains why `@eslint/js` retains `peerDependenciesMeta` with `eslint` marked as optional instead of removing the peer dependency entirely.

The optional peer dependency in `@eslint/js` is intended to signal compatibility with specific ESLint versions. It is currently `"^10.0.0"`, but if a new core rule is added in ESLint v10.X.0, the range would be bumped to `"^10.X.0"` to reflect the minimum version in which that rule exists.

I think the optional peer dependency solves a different problem and it wouldn’t help in the use case described in the OP, since npm doesn't automatically install optional peer dependencies. A mandatory peer dependency would, but that would come with other drawbacks.

fasttime (approved):
LGTM.

DMartens:
Thank you for the context lumir.
I am going to open a PR for updating the `eslint/css` documentation (`eslint/markdown` was recently updated to specify that ESLint is required).
Changes LGTM, thanks.

DMartens (approved):

PR #267, opened 2026-07-28
feat: treat hexadecimal literals as integers in `no-unsafe-values`

### LINKED ISSUE #266: Bug: `no-unsafe-values` does not report hexadecimal integers that contain the digit `e`
### Environment

```
ESLint version: v10.8.0
@eslint/json version: v2.0.1
Node version: v22.22.3
npm version: v10.8.2
Operating System: StackBlitz WebContainer (linux)
```

```
json5
```

### What did you do?

<details>

```js
// eslint.config.js

export default [
	{
		files: ["**/*.json5"],
		plugins: { json },
		language: "json/json5",
		rules: { "json/no-unsafe-values": "error" },
	},
];
```
</details>

Since a JSON5 document holds a single value, I used four files: two for the case I am reporting and two as controls.

```js
// a-hex-with-e.json5
0xe0000000000000

// d-hex-precision-loss.json5
0xdeadbeefdeadbeef

// b-hex-without-e.json5   (control)
0x20000000000000

// c-decimal.json5         (control)
9007199254740993
```

I expected all four files to be reported.

All four values are outside the safe integer range (`±2^53-1`) that this rule checks. `0xe0000000000000` is `63050394783186944` and `0xdeadbeefdeadbeef` is `16045690984833335023`, both well above `Number.MAX_SAFE_INTEGER` (`9007199254740991`). The two control values are outside that range as well, and they are reported correctly.

`0xdeadbeefdeadbeef` also loses precision when read as a JavaScript number: it becomes `16045690984833335296`, so the value silently changes.

### What actually happened?

Only the two control files are reported. Both hexadecimal values that contain the digit `e` are silently accepted:

```
❯ npx eslint .

/home/projects/stackblitz-starters-hqubvayr/b-hex-without-e.json5
  1:1  error  The integer '0x20000000000000' is outside the safe integer range  json/no-unsafe-values

/home/projects/stackblitz-starters-hqubvayr/c-decimal.json5
  1:1  error  The integer '9007199254740993' is outside the safe integer range  json/no-unsafe-values

✖ 2 problems (2 errors, 0 warnings)
```

Neither `a-hex-with-e.json5` nor `d-hex-precision-loss.json5` appears in the output at all.

```
https://stackblitz.com/edit/stackblitz-starters-hqubvayr
```

### Participation

### AI acknowledgment

### Additional comments

Hello. I looked into where this comes from, and here is what I found.

**Where the bug is**

In `src/rules/no-unsafe-values.js`, the `Number` handler decides whether a value was "intended to be an integer" by testing the raw text for `.` or `e`:

```js
// L97
} else if (!/[.e]/iu.test(value)) {
	// Intended to be an integer
```

For a hexadecimal literal, `e` and `E` are ordinary hex digits, not an exponent marker. So `0xe0000000000000` matches this test, skips the integer branch that compares against `Number.MAX_SAFE_INTEGER` (L98-L108), and falls through to the floating point branch (L109-L126) which only checks for subnormals. Nothing is reported.

This affects any hexadecimal literal containing `e` or `E` anywhere in its digits, not just ones that start with `0xe`. `0xdeadbeefdeadbeef` is an example where the value genuinely changes when read as a JavaScript number, and it is not reported either.

JSON5 defers its number grammar to ECMAScript 5.1, where a hexadecimal literal is a separate production from a decimal literal and cannot carry a fraction or an exponent part. So treating a hexadecimal literal as a floating point value is not correct in the first place.

**Relation to #194**

This is not the same as #194 (`fix: program crashes in no-unsafe-values`), even though that one also involved hexadecimal input. #194 fixed a crash in the `node.value === 0` branch, and added valid-case tests for zero-valued hex such as `0x0`, `0X0` and `+0x0`. The value here is not zero, so it never reaches that branch. It reaches the check at L97, which #194 did not change.

I could not find an existing issue or pull request covering this, so I am reporting it.

If this is accepted, I'd like to take it on.

electrohyun:
#### Prerequisites checklist

#### AI acknowledgment

Bug fix.

image
image

`no-unsafe-values` decides whether a number was intended to be an integer by testing its raw text for `.` or `e`. In a hexadecimal literal, `e` and `E` are digits rather than an exponent marker, so any hexadecimal literal containing one of them was routed to the floating point branch, which only checks for subnormals. Those values were never compared against the safe integer range and nothing was reported.

Changes:

- Added a `HEX` pattern next to the existing `NUMBER` and `NON_ZERO` constants.
- The integer branch now also accepts hexadecimal literals, so they are always compared against `Number.MAX_SAFE_INTEGER` and `Number.MIN_SAFE_INTEGER`.
- Extended the existing comment on that branch to say why hexadecimal is treated separately.

The source change is small: one new constant and one condition in the integer branch. Decimal notation is untouched: `1e10`, `1.5` and `0.00001e-10` still take the floating point branch exactly as before.

New test cases:

Invalid, previously not reported:

- `0xe0000000000000` (`63050394783186944`)
- `0XE0000000000000` (uppercase `E`)
- `0xdeadbeefdeadbeef` (`16045690984833335023`, which becomes `16045690984833335296` when read as a number)
- `+0xe0000000000000` and `-0xe0000000000000`

Valid, must stay unreported:

- `0xDEADBEEF` (`3735928559`, inside the safe integer range)

I checked that the new tests actually guard the change by reverting the source edit and running the suite again: exactly those five invalid cases fail without the fix and pass with it. The zero-valued hexadecimal cases added in #194 (`0x0`, `0X0`, `+0x0`, `+0X0`, `-0x0`, `-0X0`) are unaffected, since they are handled by the earlier `node.value === 0` branch.

`npm test` and `npm run lint` both pass.

#### Related Issues

fixes #266

Two things about the test cases:

1. The `0xDEADBEEF` valid case is there on purpose. It is a hexadecimal literal that contains `E` but is inside the safe integer range, so it guards against over-correcting this into "report every hexadecimal literal".
2. `+0x…` and `-0x…` are included because #194 dealt with signed hexadecimal input, so I wanted the signed forms covered explicitly.

### DIFF
--- a/src/rules/no-unsafe-values.js
+++ b/src/rules/no-unsafe-values.js
@@ -30,6 +30,7 @@
 const NUMBER =
 	/^[+-]?(?<int>0|([1-9]\d*))?(?:\.(?<frac>\d*))?(?:e[+-]?\d+)?$/iu;
 const NON_ZERO = /[1-9]/u;
+const HEX = /^[+-]?0x/iu;
 
 //-----------------------------------------------------------------------------
 // Rule Definition
@@ -94,8 +95,8 @@ export default /** @satisfies {NoUnsafeValuesRuleDefinition} */ ({
 								data: { value },
 							});
 						}
-					} else if (!/[.e]/iu.test(value)) {
-						// Intended to be an integer
+					} else if (HEX.test(value) || !/[.e]/iu.test(value)) {
+						// Intended to be an integer; in hex, `e` is a digit, not an exponent
 						if (
 							node.value > Number.MAX_SAFE_INTEGER ||
 							node.value < Number.MIN_SAFE_INTEGER
--- a/tests/rules/no-unsafe-values.test.js
+++ b/tests/rules/no-unsafe-values.test.js
@@ -63,6 +63,10 @@ ruleTester.run("no-unsafe-values", rule, {
 			language: "json/json5",
 		},
 		{
+			code: "0xDEADBEEF",
+			language: "json/json5",
+		},
+		{
 			code: ".0",
 			language: "json/json5",
 		},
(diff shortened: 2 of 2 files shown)


lumirlumir (commented):
Thanks for the PR. Since this fix will report more errors, I’m marking it as a feature per our policy.

https://github.com/eslint/eslint/blob/main/README.md?plain=1#L191

electrohyun:
@lumirlumir Removing a false negative means the rule reports more errors, so minor is the right bucket. I see the title has been updated. Thanks for flagging that.

lumirlumir (approved):
LGTM, thanks!

Since this is originally a bug fix but has been marked as a feature, I’m leaving it open for a second review.

DMartens (approved):
Changes LGTM, thank you.

PR #279, opened 2026-09-05
docs: clarify JSON5 `Infinity` and `NaN` handling in `no-unsafe-values`

### LINKED ISSUE #277: Bug: `no-unsafe-values` does not report JSON5 `Infinity` values
### Environment

ESLint version: v10.9.1
@eslint/json version: v2.1.0
Node version: v22.22.3
npm version: v10.8.2
Operating System: Ubuntu 20.04.0 LTS (StackBlitz)

json5

### What did you do?

<details>

```js

export default [
  {
    files: ["**/*.json5"],
    plugins: { json },
    language: "json/json5",
    rules: { "json/no-unsafe-values": "error" },
  },
];
```

</details>

```json5
[2e308, +2e308, -2e308, Infinity, +Infinity, -Infinity]
```

All six values should be reported.

### What actually happened?

Image

Only 2e308, +2e308, and -2e308 are reported.

https://stackblitz.com/edit/stackblitz-starters-jgyzylpm?file=index.js

### Participation

### AI acknowledgment

I used an AI to correct the grammar in the section below to make it clearer.

### Additional comments

This is something I thought about while working on #266, and after looking into it, it appears to be a bug. If you agree, I’d be happy to submit a PR right away.

<hr />

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

electrohyun:
#### Prerequisites checklist

#### AI acknowledgment

Fix `no-unsafe-values` not reporting JSON5 `Infinity` values.

Added an `Infinity` visitor and regression tests for `Infinity` and `-Infinity`.

#### Related Issues

fixes #277 

image

`+Infinity` is intentionally omitted to match the existing `2e308`/`-2e308` coverage (which does not include `+2e308`).

<hr />

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

## Summary by CodeRabbit

- **Documentation**
  - Clarified that explicit JSON5 `Infinity` and `NaN` values are accepted by the `no-unsafe-values` rule.
  - Documented that numeric literals exceeding JavaScript’s number range remain subject to unsafe-value checks, even when they evaluate to `Infinity`.

- **Tests**
  - Added coverage confirming valid handling of signed and unsigned `Infinity` and `NaN` JSON5 literals.
  - Removed outdated expectations that reported explicit `Infinity` values as unsafe.

### DIFF
--- a/src/rules/no-unsafe-values.js
+++ b/src/rules/no-unsafe-values.js
@@ -127,6 +127,15 @@ export default /** @satisfies {NoUnsafeValuesRuleDefinition} */ ({
 					}
 				}
 			},
+			Infinity(node) {
+				const value = context.sourceCode.getText(node);
+
+				context.report({
+					loc: node.loc,
+					messageId: "unsafeNumber",
+					data: { value },
+				});
+			},
 			String(node) {
 				if (node.value.isWellFormed) {
 					if (node.value.isWellFormed()) {
--- a/tests/rules/no-unsafe-values.test.js
+++ b/tests/rules/no-unsafe-values.test.js
@@ -107,6 +107,38 @@ ruleTester.run("no-unsafe-values", rule, {
 			],
 		},
 		{
+			code: "Infinity",
+			language: "json/json5",
+			errors: [
+				{
+					messageId: "unsafeNumber",
+					data: {
+						value: "Infinity",
+					},
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 9,
+				},
+			],
+		},
+		{
+			code: "-Infinity",
+			language: "json/json5",
+			errors: [
+				{
+					messageId: "unsafeNumber",
+					data: {
+						value: "-Infinity",
+					},
+					line: 1,
+					column: 1,
+					endLine: 1,
+					endColumn: 10,
+				},
+			],
+		},
+		{
 			code: '"\ud83d"',
 			errors: [
 				{


electrohyun:
Since this change reports more problems, I changed the prefix from `fix` to `feat`.

lumirlumir (changes requested):
> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca): confirmed.

Thanks for the PR, but I am not sure whether reporting `Infinity` is the right behavior for this rule.

`Infinity` and `NaN` are valid values according to the JSON5 specification:

https://spec.json5.org/#summary-of-features

Image

The above values are parsed as intended by the `json5` parser:

https://github.com/json5/json5/blob/main/test/parse.js#L170-L184

Image

---

After this change, the rule message would be `The number 'Infinity' will evaluate to Infinity.`, but reporting an error for using an `Infinity` value that the specification explicitly allows doesn't seem to make sense to me.

For example, `1e400` overflows to `Infinity` when parsed as a JavaScript `Number`, whereas an explicit JSON5 `Infinity` literal directly represents infinity. I think these two cases should be considered separately.

A standard JSON parser already throws an error when it encounters `Infinity`, so this additional check only applies to JSON5.

```js
JSON.parse("Infinity"); // SyntaxError
JSON.parse(Infinity); // SyntaxError
JSON.parse("1e400"); // Infinity
```

So, I think adding test cases for `Infinity` and its variants to `valid` would make more sense, to ensure that not reporting these values in JSON5 mode is intentional.

electrohyun:
Thank you for the review, @lumirlumir!

As you pointed out, an explicit JSON5 Infinity should be considered separately from a numeric literal that evaluates to `Infinity` due to overflow. I also agree that the current `unsafeNumber` message would not be appropriate in this case.

For this PR, I’ll remove the newly added reporting logic and instead add `Infinity`, `+Infinity`, and `-Infinity` as valid regression test cases.

<hr />

Separately, I have one question to make sure I understand the intended scope of the rule. 

> Does interchange here refer only to environments that support the same JSON format, such as JSON5?

When I opened the issue, I interpreted this as a broader interchange concern because an explicit Infinity is valid in JSON5 but cannot be represented in standard JSON while preserving the same meaning.

If the rule is designed to evaluate safety within each dialect, I think it may be helpful to clarify in the documentation the distinction between an explicit Infinity and a numeric literal that evaluates to Infinity due to overflow.

Would it be helpful to discuss this in a separate issue?

<hr />

Thank you again for pointing out the distinction I missed. ~I’ll update the PR by September 7.~

DMartens:
From the rule documentation:
> JSON is widely used for data interchange between systems, but certain values can cause interoperability issues when transferred between different parsers and environments

I interpreted this as any value which could cause interoperability issues.
For example C does not have a standard concept of `Infinity`.
The safe way would be to add an option for this.

electrohyun:
Thanks for the review, @DMartens!

I think it would be better to clarify the intended behavior before updating the implementation. I’ll hold off on the changes for now and wait for further thoughts.

lumirlumir (commented):
> I interpreted this as any value which could cause interoperability issues.
For example C does not have a standard concept of Infinity.
The safe way would be to add an option for this.

After reviewing the current C libraries and Rust implementation, I found that all of them support positive and negative `Infinity` and `NaN`.

Rust:

```rs
let positive = f64::INFINITY;
let negative = f64::NEG_INFINITY;
let nan = f64::NAN;

assert!(positive.is_infinite());
assert!(nan.is_nan());
``` 

C99:

```c
#include <math.h>

int main(void) {
    float positive = INFINITY;
    float negative = -INFINITY;

    int result = isinf(positive);
    int negative_result = isinf(negative);

    return 0;
}
``` 

Also, the C JSON library `yyjson` supports positive and negative `Infinity` and `NaN`.

https://github.com/ibireme/yyjson/blob/master/doc/API.md#yyjson_read_json5

image

So, I think interoperability issues with JSON5 data containing `Infinity` or `NaN` only arise when people use an inappropriate parser to exchange the data.

(As far as I know, Rust’s `f64` and C’s `double` (on IEEE 754 implementations) can also represent infinity, just like JavaScript’s `Number`: https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-language-types-number-type)

---

> If the rule is designed to evaluate safety within each dialect, I think it may be helpful to clarify in the documentation the distinction between an explicit Infinity and a numeric literal that evaluates to Infinity due to overflow.

So, I think it would be helpful to mention this rule’s limitations in the documentation and tests, making it clear that explicitly reporting `Infinity` and `NaN` is outside the scope of this rule and that an appropriate JSON5 parser will handle these values correctly, even when implemented in another language such as C or Rust.

nzakas:
I don't think we should add `Infinity` to this rule. As already noted, the purpose of the rule is to flag values that may not be interpreted correctly in different implementations. `Infinity` is a formal part of the JSON5 specification and therefore isn't unsafe.

electrohyun:
Based on the comments, I removed the `Infinity` visitor and added `Infinity`, `+Infinity`, `-Infinity`, `NaN`, `+NaN`, and `-NaN` as valid test cases. 

I also added a short explanation to the rule documentation to distinguish explicit JSON5 `Infinity` and `NaN` values from numeric literals such as `1e400` that evaluate to `Infinity`.

This is not meant to be the final version. I put it together to provide something concrete to review, and I’d appreciate any feedback, particularly on the documentation wording and PR title.

docs/rules/no-unsafe-values.md
```diff
@@ -18,6 +18,8 @@ These issues can lead to data corruption, silent failures, or inconsistent behav
 
 This rule warns on values that are unsafe for interchange, such as strings with unmatched [surrogates](https://en.wikipedia.org/wiki/UTF-16), numbers that evaluate to Infinity, numbers that evaluate to zero unintentionally, numbers that look like integers but are too large, and [subnormal numbers](https://en.wikipedia.org/wiki/Subnormal_number).
 
+For JSON5, this rule does not report explicit `Infinity` and `NaN` values because the specification defines how they are interpreted. These values differ from numeric literals such as `1e400`, which evaluate to `Infinity` because they exceed JavaScript's number range.
```
lumirlumir:
```suggestion
NOTE: For JSON5, this rule does not report explicit [`Infinity` and `NaN`](https://spec.json5.org/#summary-of-features-numbers) values because the specification defines how they are interpreted. These values differ from numeric literals such as `1e400`, which evaluate to `Infinity` because they exceed JavaScript's number range.
```

Just a small suggestion to make the documentation clearer, since we’re discussing the case that is intentionally not reported and linking to the specification. Otherwise, LGTM.

lumirlumir (changes requested):

lumirlumir (approved):
LGTM, thanks!

I’m leaving it open so others can verify the change as well.

electrohyun:
Hi! Thank you for the review.

I updated the PR description to clarify the scope of my AI use and am noting the update here as well. My earlier disclosures focused on AI material included in the submitted work. This update also documents AI assistance that did not generate the submitted changes.

For this PR, I wrote the review responses and made the technical decisions myself. I did use AI to check the English grammar in those responses, so I have included that in the disclosure. Thanks again!

docs/rules/no-unsafe-values.md
```diff
@@ -18,6 +18,8 @@ These issues can lead to data corruption, silent failures, or inconsistent behav
 
 This rule warns on values that are unsafe for interchange, such as strings with unmatched [surrogates](https://en.wikipedia.org/wiki/UTF-16), numbers that evaluate to Infinity, numbers that evaluate to zero unintentionally, numbers that look like integers but are too large, and [subnormal numbers](https://en.wikipedia.org/wiki/Subnormal_number).
 
+NOTE: For JSON5, this rule does not report explicit [`Infinity` and `NaN`](https://spec.json5.org/#summary-of-features-numbers) values because the specification defines how they are interpreted. These values differ from numeric literals such as `1e400`, which evaluate to `Infinity` because they exceed JavaScript's number range.
```
nzakas:
```suggestion
When run on JSON5 code, this rule does not report explicit [`Infinity` and `NaN`](https://spec.json5.org/#summary-of-features-numbers) values because the specification defines how they are interpreted. These values differ from numeric literals such as `1e400`, which evaluate to `Infinity` because they exceed JavaScript's number range.
```

nzakas (approved):
LGTM. Thanks.

lumirlumir (approved):
LGTM, thanks!

PR #259, opened 2026-07-15
fix: handle empty and whitespace-only JSON inputs

lumirlumir:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

#### AI acknowledgment

- [x] I did _not_ use AI to generate this PR.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

#### What is the purpose of this pull request?

This PR fixes an issue metioned in https://github.com/eslint/json/issues/249.

#### What changes did you make? (Give an overview)

I've simply bumped the `@humanwhocodes/momoa`'s version to the latest, and add test cases to verify them.

#### Related Issues

Fixes: #249

#### Is there anything you'd like reviewers to focus on?

Previously blocked by:

- [x] https://github.com/humanwhocodes/momoa/issues/235
- [x] https://github.com/humanwhocodes/momoa/issues/242
- [x] https://github.com/eslint/json/pull/281

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of empty and whitespace-only input across JSON, JSONC, and JSON5 parsing modes.
  - Invalid empty input now consistently returns a clear “Unexpected end of input found” error with accurate location details.

PR #281, opened 2026-09-05
fix: preserve parsing error locations in Bun

lumirlumir:
## Prerequisites checklist

## AI acknowledgment

### Environment

- ESLint version: 10.7.0
- @eslint/json version: 2.1.0
- npm version: 11.16.0
- Bun version: 1.4.2
- Operating System: Windows x64

json

### What did you do?

Install `eslint@10.7.0` and `@eslint/json@2.1.0`, then save this as `repro.mjs`:

```js
import { Linter } from "eslint";

const messages = new Linter().verify(
	"{\n//test\n}",
	[{ files: ["**/*.json"], plugins: { json }, language: "json/json" }],
	{ filename: "test.json" },
);
console.log(JSON.stringify(messages, null, 2));
```

Run `bun repro.mjs` and compare with `node repro.mjs`.

The parsing error includes `line: 2` and `column: 1`, as it does in Node.js.

### What actually happened?

Bun omits both properties:

```json
[
	{
		"ruleId": null,
		"fatal": true,
		"severity": 2,
		"message": "Parsing error: Unexpected character '/' found."
	}
]
```

The standalone reproduction is included above. 

Related reproduction in CI: https://github.com/eslint/json/actions/runs/33971265538/job/101320084273?pr=281

In Bun, the parser error's `line` and `column` are non-enumerable, so `...ex` in `JSONLanguage.parse()` drops them. Explicitly copying `line: ex.line` and `column: ex.column` preserves the location.

## Related Issues

Ref: https://github.com/eslint/json/pull/259

I found it while working on https://github.com/eslint/json/pull/259.

## Summary by CodeRabbit

- **Bug Fixes**
  - JSON parsing errors now report the accurate line and column where the issue occurred.

- **Tests**
  - Added coverage to verify error locations for invalid comments and trailing commas in JSON and JSONC modes.

### DIFF
--- a/src/languages/json-language.js
+++ b/src/languages/json-language.js
@@ -143,6 +143,8 @@ export class JSONLanguage {
 				errors: [
 					{
 						...ex,
+						line: ex.line,
+						column: ex.column,
 						message,
 					},
 				],
--- a/tests/languages/json-language.test.js
+++ b/tests/languages/json-language.test.js
@@ -85,6 +85,8 @@ describe("JSONLanguage", () => {
 				result.errors[0].message,
 				"Unexpected character '/' found.",
 			);
+			assert.strictEqual(result.errors[0].line, 2);
+			assert.strictEqual(result.errors[0].column, 1);
 		});
 
 		it("should not parse trailing commas by default in json mode", () => {
@@ -99,6 +101,8 @@ describe("JSONLanguage", () => {
 				result.errors[0].message,
 				"Unexpected token RBrace found.",
 			);
+			assert.strictEqual(result.errors[0].line, 3);
+			assert.strictEqual(result.errors[0].column, 1);
 		});
 
 		it("should not parse trailing commas by default in jsonc mode", () => {
@@ -113,6 +117,8 @@ describe("JSONLanguage", () => {
 				result.errors[0].message,
 				"Unexpected token RBrace found.",
 			);
+			assert.strictEqual(result.errors[0].line, 3);
+			assert.strictEqual(result.errors[0].column, 1);
 		});
 
 		it("should parse trailing commas when enabled in jsonc mode", () => {


mdjermanovic:
> In Bun, the parser error's `line` and `column` are non-enumerable

How did these properties become non-enumerable? I couldn't reproduce this locally.

DMartens:
I was able to reproduce the missing `line` and `column` property with Bun 1.4.
As lumir said, the reason is that they are not enumerable (checked via `Object.getOwnPropertyDescriptors(error)`).

mdjermanovic:
> As lumir said, the reason is that they are not enumerable

I get it, but why are they not enumerable in bun? If these properties are generated by the parser, perhaps this problem should be fixed there.

DMartens:
I could not find an issue for these properties in the Bun repository.
As the properties `line` and `column` of `Error` are non-standard, I think we should rather fix it here.

lumirlumir:
I’ve looked into the context further, and it seems that `line` and `column` properties on `Error` are non-standard and aren’t specified by ECMAScript. Therefore, whether they are enumerable may depend on the runtime implementation.

https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-error-objects

It also says:

https://tc39.es/ecma262/multipage/conformance.html#sec-conformance

image

mdjermanovic:
Okay, so Bun always sets the `line` and `column` properties on error objects. These properties represent the line and column in the file where the error was thrown, and they are non-enumerable. The problem is that we are using the same properties in the parser API, where they represent the line and column where invalid syntax was detected in the given source code. The parser overwrites values with ones expected by the API, but the properties stay non-enumerable. Is this a correct description of the cause of this problem?

lumirlumir:
> Okay, so Bun always sets the line and column properties on error objects. These properties represent the line and column in the file where the error was thrown, and they are non-enumerable. The problem is that we are using the same properties in the parser API, where they represent the line and column where invalid syntax was detected in the given source code. The parser overwrites values with ones expected by the API, but the properties stay non-enumerable. Is this a correct description of the cause of this problem?

Yes. If I understand it correctly, I believe this is the root cause of the problem.

mdjermanovic:
I can reproduce the problem now. I was running `bun` without the `--bun` flag, so it was running ESLint with Node.js I believe.

mdjermanovic:
Then, I'm not sure what the best course of action would be out of these:

1) Revisit the parser API, since properties clash with Bun's.
2) Update the parser to make properties enumerable and thus consistent across runtimes.
3) Just make the fix proposed in this PR.

@nzakas what do you think?

nzakas:
I think we can just accept this PR. Bun has a bunch of incompatibilities with Node.js, and I don't think we need to change our design to accommodate them. `line` and `column` are fairly common properties on errors thrown by parsers, so this is more than an ESLint concern. Plus, I think in most cases whether or not `line` and `column` are enumerable makes no difference at all. Just a quirk we've run into.

nzakas (approved):

mdjermanovic (approved):
LGTM, thanks!

DMartens (approved):
Changes LGTM, thanks.

PR #283, opened 2026-09-08
fix: escape normalized keys in `no-unnormalized-keys` autofix

### LINKED ISSUE #282: Bug: `no-unnormalized-keys` autofix normalizes key without escaping `"` or `\`
### Environment

ESLint version: 10.10.0
@eslint/json version: 2.1.0
Node version: 22.16.0
npm version: 11.17.0
Operating System: darwin 25.5.0

json

### What did you do?

<details>

```js

export default [
	{
		files: ["**/*.json"],
		plugins: { json },
		language: "json/json",
		rules: {
			"json/no-unnormalized-keys": ["error", { form: "NFKC" }],
		},
	},
];
```

</details>

```json
{
	"＂": 1,
	"＼": 2
}

```

Image

(Before running `npx eslint test.json --fix`)

The autofix should produce valid JSON `{"\"": 1, "\\": 2}`.

### What actually happened?

Image

(After running `npx eslint test.json --fix`)

The autofix produces invalid JSON `{""": 1, "\": 2}`.

https://stackblitz.com/edit/stackblitz-starters-vvepisbw?file=test.json (You can reproduce by running `npx eslint test.json --fix`)

### Participation

### AI acknowledgment

I used AI to check the English grammar in the issue description I wrote myself. The technical decisions in those responses were my own.

For context, Microsoft Japanese IME supports full-width punctuation, so text such as ＂りんご＂ or ＂ａｂｃ＂ can be entered. I had encountered this form while using Japanese input before, which prompted me to test the full-width quotation mark here.

### Additional comments

From my reading of the rule’s implementation and documentation, `no-unnormalized-keys` rule is intended to catch cases where different Unicode representations could cause key lookup or comparison mismatches. 

However, it now includes an autofix that normalizes object keys according to the `NFC`, `NFD`, `NFKC`, or `NFKD`, so I opened this issue because I understood converting `＂` to `"` to be within the intended scope of that autofix.

These characters may occur in practice through full-width input modes and conversions in Japanese and Chinese IMEs, so I think the autofix should preserve valid JSON when normalizing them.

<hr />

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

electrohyun:
#### Prerequisites checklist

#### AI acknowledgment

Fix `no-unnormalized-keys` autofix producing invalid JSON when normalization introduces quotes or `\`.

Escape `\` and matching quotes in normalized keys.

Add regression tests for `NFKC` and `NFKD` normalization across JSON, JSONC, and JSON5 (single quote cases are included only in JSON5).

#### Related Issues

fixes #282 

At first, I wrote code with a `let` variable, but it ended up using that variable’s name instead of `normalizedKey`, and I think this is less consistent with the rule name `no-unnormalized-keys`.

The current approach does the same by using a different name, `escapedKey`, but only for string keys and `normalizedKey` is still used for identifier keys, so I went with this structure.

If there’s a better way to structure the fixer, please let me know.

<hr />

Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic fixes for unnormalized string keys by preserving their original quote style and correctly escaping quotation marks and backslashes.
  - Ensured normalized keys are inserted safely across supported key formats.

- **Tests**
  - Expanded coverage for full-width quotation marks, apostrophes, and backslashes using Unicode escape sequences.
  - Preserved validation across supported normalization forms and JSON-derived formats.


tests/rules/no-unnormalized-keys.test.js
```diff
@@ -260,6 +260,132 @@ ruleTester.run("no-unnormalized-keys", rule, {
 				},
 			],
 		},
+		{
+			code: `{"a＂b＼c＇＂＼": 1}`,
+			output: `{"a\\"b\\\\c'\\"\\\\": 1}`,
```
electrohyun: Should we use `String.raw` for readability here?

src/rules/no-unnormalized-keys.js
```diff
@@ -85,12 +85,20 @@ export default /** @satisfies {NoUnnormalizedKeysRuleDefinition} */ ({
 								return null;
 							}
 
-							return fixer.replaceTextRange(
-								name.type === "String"
-									? [name.range[0] + 1, name.range[1] - 1]
-									: name.range,
-								normalizedKey,
-							);
+							if (name.type === "String") {
```
DMartens: We could simplify this as only the "key" is changed:
```js
const fixedKey = name.type === "String" ? escapeKey(...) : normalizedKey;
return fixer.replaceText(name, fixedKey);
```
and extract the quote + `replaceAll` logic into a function (which now adds the same quotes around the key as the whole key is replaced).
As you mentioned it is hard to come up with a better name as `normalizedKey` is in scope.

tests/rules/no-unnormalized-keys.test.js
```diff
@@ -260,6 +260,132 @@ ruleTester.run("no-unnormalized-keys", rule, {
 				},
 			],
 		},
+		{
+			code: `{"a＂b＼c＇＂＼": 1}`,
```
DMartens, line 341: Can you please add separate test cases for one `options.form` for the quotes and escaping logic (e.g. { "a'b": 1 } showing that the not used quote is not escaped).
You can keep the "combined" test cases.

DMartens (changes requested):
Thank you for the PR.
The logic looks good to me, just one suggested refactoring and additional test cases.

tests/rules/no-unnormalized-keys.test.js
DMartens: I agree that using `String.raw` for every `output` property would be more readable.

electrohyun:
@DMartens Applied. I think `fixedKey` looks good. Thank you for the review!

tests/rules/no-unnormalized-keys.test.js
```diff
@@ -260,6 +260,209 @@ ruleTester.run("no-unnormalized-keys", rule, {
 				},
 			],
 		},
+		{
+			code: `{"a＂b": 1}`,
```
lumirlumir, line 264:
```suggestion
			code: `{"a\uff02b": 1}`,
```

> Disclosure: I'm a participant of [open source contribution program OSSCA](https://github.com/eslint-ossca): confirmed.

If using U+FF02 is the intention of this test, could we use the Unicode escape sequence instead? In some code editors or IDEs, it may look similar to U+0022, which could confuse people when reading the test cases.

It would also be helpful to update the other test cases affected by this.

src/rules/no-unnormalized-keys.js
```diff
@@ -85,12 +104,15 @@ export default /** @satisfies {NoUnnormalizedKeysRuleDefinition} */ ({
...
-							return fixer.replaceTextRange(
+							const fixedKey =
 								name.type === "String"
-									? [name.range[0] + 1, name.range[1] - 1]
-									: name.range,
-								normalizedKey,
-							);
+									? escapeKey(
+											normalizedKey,
+											context.sourceCode.getText(name),
```
lumirlumir, line 111:
```suggestion
											sourceCode.getText(name),
```

```diff
+ const { sourceCode } = context;
const [{ form }] = context.options;
```

Small suggestion: I think we could destructure `sourceCode` at the top to avoid the minor overhead of accessing it each time, since there’s one more occurrence of `context.sourceCode`.

lumirlumir, line 111: I am not sure we need to call the `getText` method here. We can simply access the quote information as follows:

```js
							const fixedKey =
								name.type === "String"
									? escapeKey(
											normalizedKey,
											context.sourceCode.text[
												name.range[0]
											],
										)
									: normalizedKey;
```

The helper function could then be simplified as follows:

```js
/**
 * Escapes a normalized string key and wraps it in its original quotes.
 * @param {string} normalizedKey The normalized key to escape.
 * @param {string} quote The quote character used in the original key.
 * @returns {string} The escaped and quoted key.
 */
function escapeKey(normalizedKey, quote) {
	const escapedKey = normalizedKey
		.replaceAll("\\", "\\\\")
		.replaceAll(quote, `\\${quote}`);

	return `${quote}${escapedKey}${quote}`;
}
```

lumirlumir (changes requested):

electrohyun:
@lumirlumir Applied. Thank you for suggesting this structure!

DMartens (dismissed):
Changes LGTM, thanks. Leaving open for lumir.

src/rules/no-unnormalized-keys.js
```diff
@@ -85,12 +104,15 @@ export default /** @satisfies {NoUnnormalizedKeysRuleDefinition} */ ({
...
-									: name.range,
-								normalizedKey,
-							);
+									? escapeKey(
+											normalizedKey,
+											sourceCode.text[name.range[0]],
+										)
+									: normalizedKey;
+
+							return fixer.replaceText(name, fixedKey);
```
lumirlumir, line 115: The behavior would be the same, but rather than reconstructing the key by concatenating strings based on the `quote` information, I think we can simply escape the normalized key here.

To me, the original `type === "String" ? [range[0] + 1, range[1] - 1] : range` more clearly represents the original intent and makes it clear exactly where this fix applies.

```js
/**
 * Escapes a normalized string key for use inside its original quotes.
 * @param {string} normalizedKey The normalized key to escape.
 * @param {string} quote The quote character used in the original key.
 * @returns {string} The escaped and quoted key.
 */
function escapeKey(normalizedKey, quote) {
	return normalizedKey
		.replaceAll("\\", "\\\\")
		.replaceAll(quote, `\\${quote}`);
}
```

```js
const { loc, range, type } = node.name;
```

```js
							return fixer.replaceTextRange(
								type === "String"
									? [range[0] + 1, range[1] - 1]
									: range,
								type === "String"
									? escapeKey(
											normalizedKey,
											sourceCode.text[range[0]],
										)
									: normalizedKey,
							);
```

electrohyun:
Destructuring `node.name` and simplifying the helper function look good too.

I’ll keep the original approach while retaining the helper extraction suggested in the earlier review. Thanks for another review!

DMartens (approved):
Changes LGTM, thanks. Leaving open for lumir.

lumirlumir (approved):
LGTM, thanks!

